PR #12, opened 2015-02-05
ANY23-248 NTriplesWriter on hadoop : issue with MIME type/Upgrade sesame dependencies to 2.7.14

lewismc:
Trivial upgrade of sesame deps


lewismc:
@ansell can you take a look? Thank you Peter

PR #19, opened 2015-10-29
Fix for a ThreadSafety issue in ItemPropValue

antheque:
When multiple HTML documents are parsed concurrently in different threads, the MicrodataParser will sometimes throw very weird Exceptions or deliver broken values of date properties. I tracked it to a static SimpleDateFormat field in ItemPropValue. SimpleDateFormat is not thread-safe and should never be used like this. 

I wrote a unit test that tries to parse the same document in 10 concurrent threads about 100 times each. Then it gets a value of the "birthday" property. In all thousand cases the value should be the same. On my machine this test fails each time with various errors. With the fix - the test passes each time. 

The test uses a cyclic barrier and ensures that only in-memory data is used, so that I/O overhead and thread creation overhead do not interfere with the actual processing under test.

### DIFF
--- a/core/src/test/java/org/apache/any23/extractor/microdata/MicrodataParserTest.java
+++ b/core/src/test/java/org/apache/any23/extractor/microdata/MicrodataParserTest.java
@@ -17,21 +17,29 @@
 
 package org.apache.any23.extractor.microdata;
 
-import org.apache.any23.extractor.html.TagSoupParser;
-import org.apache.any23.util.StreamUtils;
-import org.junit.Assert;
-import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.w3c.dom.Document;
-
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.text.ParseException;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/any23/extractor/microdata/MicrodataParserTest.java
```diff
@@ -88,7 +96,58 @@ public void testGetContentAsDate() throws IOException, ParseException {
...
     }
+    
+    @Test
+    public void testGetDateConcurrent() throws IOException, ParseException {
+        final Date expectedDate = new GregorianCalendar(2009, Calendar.MAY, 10).getTime(); // 2009-05-10
+        final byte [] content = IOUtils.toByteArray(getClass().getResourceAsStream("/microdata/microdata-basic.html"));
+        final int threadCount = 10;
+        final int attemptCount = 100;
+        final List<Thread> threads = new ArrayList<Thread>();
+        final CyclicBarrier barrier = new CyclicBarrier(threadCount + 1);
```
ansell: I have typically done these tests using two separate CountDownLatch objects. Ie, a starting latch/barrier with value 1, and another latch with the number of threads as its value.

ansell:
This looks good functionally, I will test it today, and if it works I will add another commit to use CountDownLatch so it is slightly more maintainable/readable.

PR #33, opened 2017-02-16
ANY23-304 Add extractor for OpenIE 1st pass

lewismc:
This issue is a first pass at addressing https://issues.apache.org/jira/browse/ANY23-304. There is more work to be done on the new extractor. The new dependency which has been introduced has a significant impact on the size of Any23 as there are loads of new transitive dependencies. @Yongyao this is for context.

### DIFF
--- a/core/src/main/java/org/apache/any23/extractor/openie/OpenIEExtractor.java
+++ b/core/src/main/java/org/apache/any23/extractor/openie/OpenIEExtractor.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/apache/any23/rdf/RDFUtils.java
```diff
@@ -564,6 +569,4 @@ public static boolean isAbsoluteIRI(String href) {
         }
     }
 
-    private RDFUtils() {}
```
ansell: Is this signalling an intention to change the scope of this class to include non-static methods?

core/src/main/java/org/apache/any23/extractor/openie/OpenIEExtractor.java
```diff
@@ -0,0 +1,111 @@
...
+            .append(instance.extr().arg1().text())
+            .append('\t')
+            .append(instance.extr().rel().text())
+            .append('\t');
+
+            List<Argument> listArg2s = JavaConversions.seqAsJavaList(instance.extr().arg2s());
+            for(Argument argument : listArg2s) {
+                sb.append(argument.text()).append("; ");
+            }
+            System.out.println(sb.toString());
```
ansell: Presuming this is stub code right now that will eventually send results to the ExtractorResult.

core/src/main/java/org/apache/any23/rdf/RDFUtils.java
lewismc: Possibly, I will see how the remainder of the implementation goes.

core/src/main/java/org/apache/any23/extractor/openie/OpenIEExtractor.java
lewismc: Yes, correct.

lewismc:
> If there are a large number of new transitive dependencies, possibly useful to add this as a plugin rather than into the core module.

Yes I think this is an excellent suggestion. I'm going to go ahead and refactor @ansell

PR #37, opened 2017-04-11
ANY23-282 Replacement for all Sindice namespaces and URI's

lewismc:
This issue addresses https://issues.apache.org/jira/browse/ANY23-282


lewismc:
This currently results in some extraction errors/exceptions so we need to find new IRI's for the sindice entries.

lewismc:
I pushed another commit which I think better aligns the popular prefixes with the current state of affairs. I'll need to work on some of the tests which are failing... I'll do that right now.

lewismc:
any comments here folks?

lewismc:
Thanks @ansell can you please scope out the edit (regression???) I had to make to service/src/test/java/org/apache/any23/servlet/ServletTest.java
Thanks

ansell:
Never seen "rdf1" in an auto-generated document before. Its not illegal, as long as the namespace is declared correctly, just a bit strange. Having said that, it is also potentially not interoperable with broken non-namespace-aware parsers (as angle-bracket-allergic devs are prone to implement). Probably worth looking into. Try reverting your sorting of the prefix list to put the rdf prefix first again and see if that fixes it.

lewismc:
Hi @ansell I have reverted the prefix list and committed the patch to master, if we find that this creates a nasty issue moving forward we can revert the commit.

PR #39, opened 2017-06-29
ANY23-306 Add Travic Cl build/test support 

jgrzebyta:
The aims of that feature is:

- [ ] Automatic test
- [ ] Create binaries and web page 
- [ ] Deploy binaries and web page to server(s)
- [ ] Validate links and access to the binaries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+sudo: false
```
lewismc: Can you please include an Apache license header?

lewismc:
@jgrzebyta right now this pull request only satisfies the first of the above criteria.. is this correct?
How do you propose to satisfy the remaining items?
We also have a build service which runs over at https://builds.apache.org/view/A/view/Any23/job/Any23-trunk/

jgrzebyta:
@lewismc  Thanks for pointing me out the Jenkins build service. Thus there is no sense for doubling jobs so this is really about fixing Jenkins settings.

PR #38, opened 2017-06-13
ANY23-308 Adding option '-d' support for yaml file

jgrzebyta:
- Fix **ANY23-308**
  - Add unit test
  - Upadate `TikaMIMETypeDetector` class
  - Bring two utility classes into one module `apache-any23-utils`

PR #42, opened 2017-07-11
ANY23-308 [NEW] Adding option -d support for yaml files

jgrzebyta:
Continuation of #38:

- Fix **ANY23-308**
  - Add unit test
  - Upadate `TikaMIMETypeDetector` class
  - Bring two utility classes into one module `apache-any23-utils`


jgrzebyta:
Hi @ansell  @lewismc  ,

Just for testing I have added another **travis** branch at https://github.com/jgrzebyta/any23/tree/travis to run all unit tests with command `mvn -B clean test`. Tests results in displayed in README file. 

What do you think? Would it be useful to run tests on travis and build on the jenkins as it is now? Or is it better to run both on jenkins?

ansell:
Jenkins is good the way it is right now doing both testing and deploying snapshots.

Just running the usual java sequence without customizing the "script" on TravisCI is the simplest to have it as an alternate regression test site.

core/src/test/java/org/apache/any23/extractor/yaml/YAMLExtractorTest.java
```diff
@@ -93,4 +92,17 @@ public void nullTest()
...
+
+    /**
+     * Comma separated values are parsed as well.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void csvTest()
+            throws Exception {
+        assertExtract("/org/apache/any23/extractor/csv/test-comma.csv");
```
ansell: I am not familiar with YAML, but is CSV (RFC 4180 Section 2 and all other variations) a strict subset of YAML or is it incidental that this test file works with the YAML parser? If it is only incidental we shouldn't be testing it here.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+            } catch (Exception ex) {
+                //do nothing
+            }
+
+            // final break 
+            break;
+        }
+
+        is.reset(); // MUST BE AT THE END
```
ansell: If this must be called last, it should be done in a try-finally block. If this must be done at the end, it isn't clear why it is also being done inside of the loop.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+     * @param is {@link InputStream}
+     * @return
+     * @throws IOException
+     */
+    public static boolean isYAML(InputStream is) throws IOException {
+        if (is == null) {
+            return false;
+        }
+
+        if (!is.markSupported()) {
```
ansell: The original InputStream will be modified without the ability to roll it back in this case. Simply try to call is.mark and if that fails, they need to change their calling code to provide a markable InputStream and try again.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+            is = new BufferedInputStream(is);
+        }
+
+        boolean result = false;
```
ansell: A variable isn't a great idea in the default false case, as you can simply return true whenever you decide it is try (and rely on try-finally to cleanup), and then return false at the end of the method.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+                if (Iterables.size(parsedOut) > 0) {
+                    result = true;
+                    break;
+                }
+            } catch (Exception ex) {
+                //do nothing
+            }
+
+            // final break 
+            break;
```
ansell: breaking at the end of a while(true) loop with no calls to "continue" shows the code needs to be changed. There doesn't seem to be anything iterative about the algorithm, so not clear why you are doing it this way.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+            try {
+                Yaml yml = new Yaml();
+                Iterable<Object> parsedOut = yml.loadAll(is);
+
+                if (Iterables.size(parsedOut) > 0) {
+                    result = true;
+                    break;
+                }
+            } catch (Exception ex) {
+                //do nothing
```
ansell: This should return false rather than silently aborting

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
```diff
@@ -0,0 +1,105 @@
...
+        if (!is.markSupported()) {
+            is = new BufferedInputStream(is);
+        }
+
+        boolean result = false;
+
+        // mark the reading frame position. MUST BE FIRST
+        is.mark(Integer.MAX_VALUE);
+
+        while (true) {
```
ansell: This loop is not being used. Replace it with try-finally and replace the "result =..." cases with "return ...".

core/src/test/java/org/apache/any23/extractor/yaml/YAMLExtractorTest.java
jgrzebyta: That is false positive. It is just proof that for text-based formats additional recognising should be done from `meta`. I will delete that.

utils/src/main/java/org/apache/any23/extractor/yaml/YAMLValidator.java
jgrzebyta: I have checked in the source and `markSupported()` method olny contains `return true` or `return false`.

jgrzebyta:
I have changed code and yaml mime type is detected based on file name only rather by the content. I removed useless code: restored `csvutils` module.

ansell:
I am fixing some compile and test errors and merging it into master. Thanks!

PR #44, opened 2017-07-14
ANY23-310 Solved problems with wrong loggers

jgrzebyta:
- Add a basic unit test for `Rover`
- Piped correct logger in `Rover`
- Switch on logging content size
- use `StringUtils#join` for merging parses with "comma-space" delimiter


core/src/main/java/org/apache/any23/writer/LoggingTripleHandler.java
```diff
@@ -111,6 +116,6 @@ public void setContentLength(long contentLength) {
     }
 
     private void printHeader(PrintWriter writer) {
-        writer.println("# Document-IRI\tContent-Length\tElapsed-Time\tSuccess\tExtractors");
+        writer.println("# Document-IRI\tContent-Length\tElapsed-Time\tSuccess\tExtractors:Triples");
```
lewismc: @jgrzebyta is the intention here to print the number of triples extracted? If so then this is not current behavior with Rover. It could be default behaviour if we switched it to use counting triple handler... just a thought.

jgrzebyta: Hi,

Yes that is the intention. That behaviour was implemented in the `LoggingTripleHandler` and was displayed but I have just added header for that. For example in last `[ yaml:531 ]` in the logging file only. That information was not displayed in *strerr* output.

lewismc: Hi @jgrzebyta I get it, thank you for clarifying

PR #45, opened 2017-10-23
Any23-312 simplify yaml parsing outcome

jgrzebyta:
- Simplify graph
- In case if a triple is not completed (missing object) than it is not created (previous time RDF:NIL was added as object). Still predicates characteristic (`[] rdf:type rdf:predicate ; rdfs:label ?name`) is added to final graph.
- Make subgraph blank nodes based.


jgrzebyta:
Guys!  I am still not convinced that `rdf:List` is a good choice for RDF-ised a list in yaml. Manipulation on that, without loosing order, with `sparql` query is challenging I would say.

lewismc:
@jgrzebyta I am not sure about the ```ref:List``` either. Do you have examples of 
 * an rdf:List, and
 * a SPARQL query which may loose order
?

core/src/main/java/org/apache/any23/extractor/yaml/ElementsProcessor.java
```diff
@@ -0,0 +1,201 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
```
lewismc: No need to include Copyright in any new files. These quickly become out-of-date from year to year. We just put a generic license header and then in NOTICE include the Copyright year. Can you please remove the Copyright? Thanks @jgrzebyta

jgrzebyta:
@lewismc thanks for pointing the copyright issue because I found that in several files.

I will add unit tests for sparql example. But I am pro `rdf:List` because it is parsable by `org.eclipse.rdf4j.model.util.RDFCollections` into java collection.

lewismc:
Hi @jgrzebyta I pulled the code and tested it... all looks good and test successfully. I think the additions here make good improvements. I am +1 to merge.

jgrzebyta:
@lewismc OK. Thanks. First I need to check if the RDF graph is parsable to any final graph.

PR #51, opened 2017-12-26
Solved ANY23-316

jgrzebyta:
Add handler for `null` values. Parser encodes all `null` values as RDF IRI `http://yaml.org/spec/1.2/spec.html#Null`.

- create RDF value for null values: yaml:Null
- update null-aware unit test
- modify example for unit test: 'null' value is an item of a list


core/src/test/java/org/apache/any23/extractor/yaml/YAMLExtractorTest.java
```diff
@@ -125,7 +121,7 @@ public void treeTest() throws Exception {
             RDFUtils.triple(RDFUtils.bnode(), RDFUtils.iri(ns, "value3"), RDFUtils.bnode("10")),
             RDFUtils.triple(RDFUtils.bnode("10"), RDF.FIRST, RDFUtils.bnode("11")),
             RDFUtils.triple(RDFUtils.bnode("11"), RDFUtils.iri(ns, "key3.1"), RDFUtils.bnode("12")),
-            RDFUtils.triple(RDFUtils.bnode("12"), RDF.TYPE, RDF.LIST),
+            // RDFUtils.triple(RDFUtils.bnode("12"), RDF.TYPE, RDF.LIST),
```
lewismc: @jgrzebyta I would just remove this code.

PR #58, opened 2018-01-21
ANY23-324 Changed default html parser from NekoHTML to Jsoup. 

HansBrende:
This should also indirectly fix ANY23-317, ANY23-273, ANY23-267, and ANY23-326.

I also added a "any23.tagsoup.legacy" flag in default-configuration.properties to allow switching back to the NekoHTML version.

I ran the full suite of tests against this fork using "mvn clean install", and all tests are passing.

Please let me know if there's anything more I can do to improve this pull request!

PR #59, opened 2018-01-24
ANY23-326 fixed rdfa issue with unclosed input & meta tags

HansBrende:
This PR should fix ANY23-326, ANY23-317, ANY23-273, ANY23-267, ANY23-271, ANY23-227, ANY23-268, and ANY23-266.

These all have to do with the RDFa implementation failing to parse HTML.

My previous PR (#58) attempted to fix these issues by changing the default parser from NekoHTML to Jsoup. But alas, it turns out the RDFa implementation is using a completely different html parser under the hood, and it's the RDFa parser that's too strict, not ours, so changing ours from NekoHTML to Jsoup had no effect (although it did come with a nice 20% speed increase, so there's that). It seems that, for rio parsers, the input stream is parsed with Jsoup *only to get the document language* and then parsed **again** under the hood with who knows what.

Now, I simply check the RDF format to see if we're putting out XHTML. If we are, I first XHTML-ify the stream with Jsoup before sending it on to the rio RDF parser.

mvn clean install -> all tests passed.


core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
```diff
@@ -22,16 +22,20 @@
 import org.apache.any23.extractor.ExtractionParameters;
 import org.apache.any23.extractor.ExtractionResult;
 import org.apache.any23.extractor.Extractor;
-import org.eclipse.rdf4j.rio.RDFHandlerException;
-import org.eclipse.rdf4j.rio.RDFParseException;
-import org.eclipse.rdf4j.rio.RDFParser;
-import org.eclipse.rdf4j.rio.RioSetting;
+import org.eclipse.rdf4j.rio.*;
```
lewismc: Can you please make explicit imports rather than wildcard.

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
```diff
@@ -105,7 +109,24 @@ public void run(
             parser.getParserConfig().addNonFatalError(BasicParserSettings.NORMALIZE_DATATYPE_VALUES);
             //ByteBuffer seems to represent incorrect content. Need to make sure it is the content
             //of the <script> node and not anything else!
-            parser.parse(in, extractionContext.getDocumentIRI().stringValue());
+            RDFFormat format = parser.getRDFFormat();
+            String iri = extractionContext.getDocumentIRI().stringValue();
+
+            if (format.hasFileExtension("xhtml")) {
```
lewismc: What happens when the file suffix is just .html?

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
HansBrende: Sure, sorry, that was something that IntelliJ did. I have no idea why. Will fix.

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
HansBrende: All of the RDF formats defined in the RDFFormat class which have file extension "html" also have file extension "xhtml". (And there is exactly one of them: the RDFA format--which is the one that is giving us problems.)

lewismc: ack, please ignore

HansBrende: @lewismc I can also do the check via pattern matching on MIME types instead of file extensions, if you'd prefer. This way was just the easiest & quickest, and either way we'd get the same results.

HansBrende:
@lewismc my latest commit should fix ANY23-227 and ANY23-268.

HansBrende:
@lewismc I rebased all the commits into a single commit to make it easier to see what has changed. Everything should be fully functional now.

PR #62, opened 2018-01-29
ANY23-327 Change log level to debug for RDFUtils.isAbsoluteIRI()

HansBrende:
No need to see a stack trace every time `isAbsoluteIRI` returns `false`.


core/src/main/java/org/apache/any23/rdf/RDFUtils.java
```diff
@@ -528,10 +528,10 @@ public static boolean isAbsoluteIRI(String href) {
             new java.net.URI(href.trim());
```
HansBrende: On an unrelated note, why are we testing this on `href.trim()` instead of just `href`? 
`RDFUtils.iri(href)` does *not* call `href.trim()`!

lewismc:
@jgrzebyta can you merge to master?

PR #65, opened 2018-02-11
ANY23-328 Strip comments from json-ld to make parsing more lenient

HansBrende:
I stripped single-line and multi-line comments from json-ld content, as well as CDATA markers.

Also added a test case.

`mvn clean install` -> All tests pass.


HansBrende:
@lewismc think we can get this PR put into the 2.2 release as well?

lewismc:
+1, tested locally. Anyone else have comments?

HansBrende:
@lewismc I just refactored this to use the `java.io.PushbackInputStream` class instead of my own custom pushback buffer logic. Should make this commit less complex and more readable!

HansBrende:
FYI, at around 1pm CST today, I refactored and rebased this commit to use
java.io.PushbackInputStream instead of my own custom pushback logic. This
should make things less complicated and more readable. So if you were
testing this PR before 1pm, please pull the current version.

HansBrende:
@lewismc can this be pulled into master or are you still waiting on additional feedback? If I need to change anything, let me know.

PR #67, opened 2018-03-30
ANY23-339 fixes itemscope hashcode collision problem, allows absolute URIs as subjects

HansBrende:
I fixed the itemscope hashcode collision problem documented in ANY23-339, as well as loosening the restriction on subject resources to allow absolute URIs as subjects. 

mvn clean test --> all tests pass

@lewismc Any comments before I merge this?


lewismc:
I haven't read The Reality Dysfunction... but the PR look good :)

PR #68, opened 2018-03-30
ANY23-340 Removes doctypes to allow extraction of additional rdfa 1.1 triples

HansBrende:
This should resolve the remaining problems with the Nutch test cases.

mvn clean test --> all tests pass.

@lewismc any comments?


lewismc:
+1

PR #69, opened 2018-04-02
ANY23-336 Hacky patch to tide us over until jsonldjava 0.11.2 release

HansBrende:
This is a hacky patch for Any23 that does exactly the same thing my [JSONLD-JAVA pull request](https://github.com/jsonld-java/jsonld-java/pull/228) does.

With this patch in place, the amortized extraction time for the site https://www.guthriegreen.com dropped from 30 seconds to 0.3 seconds, a difference of **two orders of magnitude**!!! 😱😮😵

This patch should be removed as soon as we can upgrade to JSONLD-JAVA version 0.11.2.

mvn clean test -> all tests pass.

@lewismc @ansell Any comments before I merge this?

### DIFF
--- a/core/src/main/java/org/apache/any23/extractor/rdf/JSONLDExtractor.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdf/JSONLDExtractor.java
@@ -17,10 +17,27 @@
 
 package org.apache.any23.extractor.rdf;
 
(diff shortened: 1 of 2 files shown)


lewismc:
I pulled this locally and tested. In terms of the extraction performance, I can confirm a significant improvement. I would be +1 to merging into master.

PR #71, opened 2018-04-03
ANY23-341 Remove dependency on defunct commons-httpclient

HansBrende:
This is my first stab at removing the decade-out-of-date `commons-httpclient` dependency.

All tests are passing, but I want another set of eyes on this.

@lewismc any comments?

### DIFF
--- a/core/src/main/java/org/apache/any23/extractor/html/HTMLDocument.java
+++ b/core/src/main/java/org/apache/any23/extractor/html/HTMLDocument.java
@@ -24,6 +24,7 @@ import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/any23/extractor/html/HTMLDocument.java
```diff
@@ -375,15 +376,16 @@ public String getDefaultLanguage() {
 
     private java.net.URI getBaseIRI() throws ExtractionException {
         if (baseIRI == null) {
+            String uri = (document instanceof Document ? (Document)document : document.getOwnerDocument()).getDocumentURI();
             try {
-                if (document.getBaseURI() == null) {
-                    log.warn("document.getBaseURI() is null, this should not happen");
+                if (uri == null) {
+                    log.warn("document.getBaseURI() is null, this should not happen", new Exception());
```
lewismc: Can the Exception be more specific?

HansBrende: @lewismc Oops, yeah, I added that exception in there just to get the stacktrace for debugging purposes, but forgot to take it out again!

PR #82, opened 2018-04-30
ANY23-231: Use RDF4J writer to return JSON-LD outputs.

JulioCCBUcuenca:
User JSON-LD as a default JSONWritter.

Having the statement:
```xml
<http://sub/1> <http://pred/1> "123"^^<http://datatype> <http://graph/1>.
```

Before using the default RDF4J writer:
```json
[
  {
    "type": "uri",
    "value": "http://sub/1"
  },
  "http://pred/1",
  {
    "type": "literal",
    "value": "123",
    "lang": null,
    "datatype": "http://datatype"
  },
  "http://graph/1"
]
```

Now, using the RDF4J writer:
```json
[
  {
    "@graph": [
      {
        "@id": "http://sub/1",
        "http://pred/1": [
          {
            "@type": "http://datatype",
            "@value": "123"
          }
        ]
      }
    ],
    "@id": "http://graph/1"
  }
]
```

I am wondering about name conventions and some drawback compatibility. 
- Should we rename the classes `JSONWriter` to `JSONLDWriter` and `JSONWriterFactory` to `JSONLDWriterFactory`. 
- When using the REST API, should we use `jsonld` as output format  instead of `json`.

Closes #ANY231 [issue](https://issues.apache.org/jira/browse/ANY23-231)


lewismc:
This looks really nice @JulioCCBUcuenca I am going to suggest the following

1. Lets retain the JSONWriter... please add your code additions to a new JSONLDWriter.
1. Add a new JSONLDWriterTest.java for the same

lewismc:
@JulioCCBUcuenca 

> Do those changes aim that Any23 supports both json and json-ld outputs?

Yes, please just leave the JSONWriter as it and add the JSONLDWriter implementation.

> If so, I think further modifications will be necessary: ...

Yes please go ahead with these as well you are correct.

> In adition, if we retain the original Json writer, we should devise another strategy to deal with ANY-231, probably a wrapper as I suggested in the issue's comments. am I right?

Yes we probably should however this is not a huge deal. It _may_ actually be better if the output stream is written to one line... I suppose this is down to preference.

JulioCCBUcuenca:
@lewismc Done! How can I update the site documentation? This has to be documented [here](http://any23.apache.org) once the master branch is released, right?

lewismc:
@JulioCCBUcuenca very nice.... very nice indeed. A lot cleaner.

> How can I update the site documentation?

https://cwiki.apache.org/confluence/display/ANY23/Building+Apache+Any23+Website+HOW_TO

The Website source lives at https://github.com/apache/any23/tree/master/src/site

JulioCCBUcuenca:
The site is updated, but I don't have appropriate credentials for publishing it through the SVN repository.

lewismc:
Thats fine @JulioCCBUcuenca I will deal with it all... thank you for the patch.
Can you make sure that you are populating the reporting page on the Any23 wiki and sending me a hyperlink to your weekly report at the end of every week? Thank you

lewismc:
@JulioCCBUcuenca can you please close off the JIRA issue. Thank you :) excellent work.

PR #83, opened 2018-05-16
ANY23-352 updated to rdf4j version 2.3.2

HansBrende:

@lewismc @JulioCCBUcuenca rdf4j version 2.3.2 defaults to pretty printing in the JSONLDWriter, so this PR seems to complement ANY23-231 nicely. Does that work for you?


lewismc:
+1 nice work

PR #86, opened 2018-06-27
ANY23-351 fixed NullPointerException in HCardExtractor

HansBrende:
This PR:
1. fixes the NullPointerExceptions that have occurred in HCardExtractor
2. supports the 'srcset' attribute in obtaining urls from img elements
3. fixes the fallback url extraction method to align with the spec by only obtaining url text from 'value'-class elements, or, if none defined, from non-'type'-class elements.


HansBrende:
@lewismc any comments?

lewismc:
Pull PR and tested locally. I've inspected it and have no comments for improvement. +1 from me.

PR #88, opened 2018-06-28
ANY23-347 fixed RDFParseExceptions caused by unbound xml namespaces

HansBrende:
My solution to this problem was to strip all xml namespaces from attributes and tag names except the "xml:" and "xmlns:" namespaces. I also added a test case.

@lewismc any comments?


lewismc:
+1

PR #89, opened 2018-06-28
ANY23-353 fallback on string datatype for langstring with no lang

HansBrende:
For string literals of datatype "langString" with no language tag, I'm simply falling back to type "string" (or using the default language if specified).

@lewismc any comments?


lewismc:
+1

PR #90, opened 2018-06-29
ANY23-355: Deprecate RDFa11Parser since Rio implementations are used instead

JulioCCBUcuenca:
The [RDFParserFactory](https://github.com/apache/any23/blob/master/core/src/main/java/org/apache/any23/extractor/rdf/RDFParserFactory.java) uses Rio implementations, and as a consequence of this, the [RDFa11Parser](https://github.com/apache/any23/blob/master/core/src/main/java/org/apache/any23/extractor/rdfa/RDFa11Parser.java) is not longer used. Thus, it needs to be deprecated.


JulioCCBUcuenca:
I deprecated all the class `RDFa11Parser`. Yeah, you are right @HansBrende, I will deprecate the exception too.

JulioCCBUcuenca:
Ohhh, I always thought that using `@deprecated`, it also deprecates the class. Thanks for pointing that out.  Done! :smile:

lewismc:
+1

PR #104, opened 2018-07-19
Any23 295: Implement ability to use librdfa

JulioCCBUcuenca:
Integration of [librdfa](https://github.com/rdfa/librdfa) on Any23 through the Rio API of RDF4J. For that, I have created a [new parser](https://github.com/JulioCCBUcuenca/librdfa-rdf4j).

In order to try out the PR, you should:

1. Install [librdfa](https://github.com/rdfa/librdfa)
2. Compile [librdfa-rdf4j](https://github.com/JulioCCBUcuenca/librdfa-rdf4j) since is not available in mvn central repository yet.
3. Compile Any23

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -60,6 +60,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>apache-any23-librdfa</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+    </dependency>
     <!-- END: Any23 -->
 
     <!-- BEGIN: httpcomponents -->
--- a/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractorFactory.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractorFactory.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.any23.extractor.rdfa;
+
+import java.util.Arrays;
+import org.apache.any23.extractor.ExtractorDescription;
+import org.apache.any23.extractor.ExtractorFactory;
+import org.apache.any23.extractor.SimpleExtractorFactory;
+import org.apache.any23.rdf.Prefixes;
+
+/**
+ *
+ * @author Julio Caguano
+ */
+public class LibRdfaExtractorFactory extends SimpleExtractorFactory<LibRdfaExtractor>
+        implements ExtractorFactory<LibRdfaExtractor> {
+
+    public static final String NAME = "html-librdfa";
+    public static final Prefixes PREFIXES = null;
+
+    private static final ExtractorDescription descriptionInstance = new RDFa11ExtractorFactory();
+
+    public LibRdfaExtractorFactory() {
+        super(RDFa11ExtractorFactory.NAME,
+                RDFa11ExtractorFactory.PREFIXES,
+                Arrays.asList("text/html;q=0.3", "application/xhtml+xml;q=0.3"),
+                "example-rdfa11.html");
+    }
+
+    @Override
+    public LibRdfaExtractor createExtractor() {
+        return new LibRdfaExtractor();
+    }
+
+    public static ExtractorDescription getDescriptionInstance() {
+        return descriptionInstance;
+    }
+}
--- a/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractorFactory.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractorFactory.java
@@ -31,11 +31,11 @@ public class LibRdfaExtractorFactory extends SimpleExtractorFactory<LibRdfaExtra
     public static final String NAME = "html-librdfa";
     public static final Prefixes PREFIXES = null;
 
-    private static final ExtractorDescription descriptionInstance = new RDFa11ExtractorFactory();
+    private static final ExtractorDescription descriptionInstance = new LibRdfaExtractorFactory();
 
     public LibRdfaExtractorFactory() {
-        super(RDFa11ExtractorFactory.NAME,
-                RDFa11ExtractorFactory.PREFIXES,
+        super(LibRdfaExtractorFactory.NAME,
+                LibRdfaExtractorFactory.PREFIXES,
                 Arrays.asList("text/html;q=0.3", "application/xhtml+xml;q=0.3"),
                 "example-rdfa11.html");
     }
--- a/core/src/main/java/org/apache/any23/extractor/rdf/RDFParserFactory.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdf/RDFParserFactory.java
@@ -17,10 +17,16 @@
 
 package org.apache.any23.extractor.rdf;
 
-import org.apache.any23.extractor.IssueReport;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+import java.util.Collections;
+import java.util.HashSet;
 import org.apache.any23.extractor.ExtractionContext;
 import org.apache.any23.extractor.ExtractionResult;
+import org.apache.any23.extractor.IssueReport;
 import org.apache.any23.rdf.Any23ValueFactoryWrapper;
+import org.apache.any23.rdf.rdfa.LibrdfaRDFaParser;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.rio.ParseErrorListener;
 import org.eclipse.rdf4j.rio.RDFFormat;
@@ -35,12 +41,6 @@ import org.eclipse.rdf4j.rio.turtle.TurtleParser;
 import org.semanticweb.owlapi.rio.OWLAPIRDFFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.Reader;
-import java.util.Collections;
-import java.util.HashSet;
 
 /**
  * This factory provides a common logic for creating and configuring correctly
@@ -120,6 +120,27 @@ public class RDFParserFactory {
             final ExtractionResult extractionResult
     ) {
         final RDFParser parser = Rio.createParser(RDFFormat.RDFA);
+        parser.getParserConfig().set(RDFaParserSettings.RDFA_COMPATIBILITY, RDFaVersion.RDFA_1_1);
+        configureParser(parser, verifyDataType, stopAtFirstError, extractionContext, extractionResult);
+        return parser;
+    }
+    
+    /**
+     * Returns a new instance of a configured RDFaParser using the librdfa library.
+     *
+     * @param verifyDataType data verification enable if <code>true</code>.
+     * @param stopAtFirstError the parser stops at first error if <code>true</code>.
+     * @param extractionContext the extraction context where the parser is used.
+     * @param extractionResult the output extraction result.
+     * @return a new instance of a configured RDFXML parser.
+     */
+    public RDFParser getRDFaLibrdfaParser(
+            final boolean verifyDataType,
+            final boolean stopAtFirstError,
+            final ExtractionContext extractionContext,
+            final ExtractionResult extractionResult
+    ) {
+        final RDFParser parser = new LibrdfaRDFaParser();
         parser.getParserConfig().set(RDFaParserSettings.RDFA_COMPATIBILITY, RDFaVersion.RDFA_1_1);
         configureParser(parser, verifyDataType, stopAtFirstError, extractionContext, extractionResult);
         return parser;
--- a/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractor.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractor.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.any23.extractor.rdfa;
+
+import org.apache.any23.extractor.ExtractionContext;
+import org.apache.any23.extractor.ExtractionResult;
+import org.apache.any23.extractor.ExtractorDescription;
+import org.apache.any23.extractor.rdf.BaseRDFExtractor;
+import org.eclipse.rdf4j.rio.RDFParser;
+
+/**
+ *
+ * @author Julio Caguano
+ */
+public class LibRdfaExtractor extends BaseRDFExtractor {
+
+    public LibRdfaExtractor(boolean verifyDataType, boolean stopAtFirstError) {
+        super(verifyDataType, stopAtFirstError);
+    }
+
+    public LibRdfaExtractor() {
+        this(false, false);
+    }
+
+    @Override
+    protected RDFParser getParser(ExtractionContext extractionContext, ExtractionResult extractionResult) {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
+    }
+
+    @Override
+    public ExtractorDescription getDescription() {
+        return LibRdfaExtractorFactory.getDescriptionInstance();
+    }
+
+}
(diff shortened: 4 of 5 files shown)


api/src/main/resources/default-configuration.properties
```diff
@@ -47,6 +47,14 @@ any23.extraction.metadata.domain.per.entity=off
 # registered the RDFa 1.0 legacy one (org.deri.any23.extractor.rdfa.RDFaExtractor).
```
lewismc: There should be no mention of ```org.deri.any23``` in this file. It should be ```org.apache.any23```, can you please correct this. Thanks

api/src/main/resources/default-configuration.properties
```diff
@@ -47,6 +47,14 @@ any23.extraction.metadata.domain.per.entity=off
...
 any23.extraction.rdfa.programmatic=on
 
+# Allows to enable Librdfa Extractor.
+# If 'on' will override the extractors with the programmatic option,
+# RDFa 1.1 Extractor (org.deri.any23.extractor.rdfa.RDFa11Extractor) and
```
lewismc: Replace ```deri``` with ```apache```

core/pom.xml
```diff
@@ -62,6 +62,14 @@
     </dependency>
     <!-- END: Any23 -->
 
+    <!-- BEGIN: Librdfa-RDF4J, loading from repository -->
+    <dependency>
+        <groupId>${project.groupId}</groupId>
```
lewismc: Please use 2 space indents the same as the rest of the file.

core/pom.xml
```diff
@@ -62,6 +62,14 @@
...
+        <artifactId>apache-any23-librdfa</artifactId>
+        <version>1.0.0</version>
```
lewismc: Ensure that the versioning is consistent with the rest of the codebase e.g. 2.3-SNAPSHOT or whatever it is currently.

core/pom.xml
```diff
@@ -335,7 +343,14 @@
     </dependency>
     <!-- END: Test Dependencies -->
   </dependencies>
-
+  
+  <repositories>
+    <repository>
+      <id>librdfa-rdf4j</id>
+      <url>https://raw.github.com/JulioCCBUcuenca/librdfa-java/repository/</url>
```
lewismc: This is a bit hacky... what is meant to reside at this URL?

core/src/main/java/org/apache/any23/extractor/rdfa/LibRdfaExtractorFactory.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2018 The Apache Software Foundation.
```
lewismc: Please remove Copyright line

librdfa-rdf4j/README.MD
```diff
@@ -0,0 +1,63 @@
...
+
+You need to install the [librdfa](https://github.com/rdfa/librdfa) library.
+
+## Install
+
+``` mvn
+<dependency>
+   <groupId>org.apache.any23</groupId>
+   <artifactId>apache-any23-librdfa</artifactId>
+   <version>1.0.0</version>
```
lewismc: Make sure the versioning is inline with existing Any23 versioning.

librdfa-rdf4j/README.MD
```diff
@@ -0,0 +1,63 @@
...
+</dependency>
+
+<repositories>
+   <repository>
+       <id>librdfa-rdf4j</id>
+       <url>https://raw.github.com/JulioCCBUcuenca/librdfa-java/repository/</url>
```
lewismc: ... again, this looks quite untidy. What resides at this URL?

librdfa-rdf4j/README.MD
```diff
@@ -0,0 +1,63 @@
...
+rdfParser.setRDFHandler(new StatementCollector(model));
+rdfParser.parse(in, "http://www.example.org./");
+```
+
+## Benchmarking
+
+In general librdfa is 2-5 seconds faster than semargl.  
+
+### librdfa-rdf4j
+- round: 0.11 [+- 0.00]
```
lewismc: What is the unit of these floating points?

librdfa-rdf4j/pom.xml
```diff
@@ -0,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lewismc: Add Apache License header.

librdfa-rdf4j/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
lewismc: Make sure XML is formatted at 2 space indents to match existing XML for Any23 codebase.

librdfa-rdf4j/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.any23</groupId>
+    <artifactId>apache-any23-librdfa</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
```
lewismc: This version needs to match the master versioning for Any23.

librdfa-rdf4j/src/main/c/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 2.8)
```
lewismc: Missing license header.

librdfa-rdf4j/src/main/c/RdfaParser.cpp
```diff
@@ -0,0 +1,19 @@
+/*
```
lewismc: Is there not an ALv2.0 version of this file?

librdfa-rdf4j/src/main/c/RdfaParser.h
```diff
@@ -0,0 +1,124 @@
+/*
```
lewismc: Is there not an Apache License v2.0 version of this file?

librdfa-rdf4j/src/main/c/RdfaParser.h
```diff
@@ -0,0 +1,124 @@
...
+#include <rdfa_utils.h>
+#include <stdlib.h>
+
+struct rdfacontext;
+
+class Callback {
+public:
+
+    virtual ~Callback() {
+        //std::cout << "Callback::~Callback()" << std::endl;
```
lewismc: Please remove if not in use.

librdfa-rdf4j/src/main/c/RdfaParser.h
```diff
@@ -0,0 +1,124 @@
...
+     */
+    rdfacontext* mBaseContext;
+
+    RdfaParser(const char* baseUri) : _callback(0) {
+        mBaseUri = baseUri;
+        mBaseContext = rdfa_create_context(baseUri);
+    }
+
+    /**
+     * Standard destructor.
```
lewismc: ```constructor```

librdfa-rdf4j/src/main/c/main.java
```diff
@@ -0,0 +1,82 @@
+
```
lewismc: Missing license header.

librdfa-rdf4j/src/main/c/main.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.any23.rdf.librdfa.RdfaParser;
+
+public class main {
+
+  public static void main(String argv[]) {
+    System.loadLibrary("rdfaJava"); // Attempts to load example.dll (on Windows) or libexample.so (on Linux)
+
+    System.out.println("Adding and calling a normal C++ callback");
+    System.out.println("----------------------------------------");
+    String ds = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
lewismc: This is very messy, it is not required please remove.

librdfa-rdf4j/src/main/c/main.java
```diff
@@ -0,0 +1,82 @@
...
+        + "      xmlns:dc=\"http://purl.org/dc/elements/1.1/\">\n"
+        + "<head>\n"
+        + "	<title>Test 0001</title>\n"
+        + "<script type=\"text/javascript\" src=\"https://ff.kis.v2.scr.kaspersky-labs.com/89CF2878-190C-6147-88B9-F9B46CAB8AD0/main.js\" charset=\"UTF-8\"></script></head>\n"
+        + "<body>\n"
+        + "	<p>This photo was taken by <span class=\"author\" about=\"photo1.jpg\" property=\"dc:creator\">Mark Birbeck</span>.</p>\n"
+        + "</body>\n"
+        + "</html>";
+
+    RdfaParser caller = new RdfaParser("http://www.google.com/");
```
lewismc: This should never be hardcoded. It should be read from main arg[]... correct?

librdfa-rdf4j/src/main/c/main.java
```diff
@@ -0,0 +1,82 @@
...
+    caller.init();
+    Callback callback = new JavaCallback(new ByteArrayInputStream(ds.getBytes(StandardCharsets.UTF_8)));
+    caller.setCallback(callback);
+
+    caller.parse();
+    //rdfa.set_rdfa_parser(caller);
```
lewismc: If not in use, please remove.

librdfa-rdf4j/src/main/c/main.java
```diff
@@ -0,0 +1,82 @@
...
+  int len = 0;
+
+  public JavaCallback(InputStream is) {
+    super();
+    bis = new BufferedReader(new InputStreamReader(is));
+  }
+
+  @Override
+  public void default_graph(String subject, String predicate, String object, int object_type, String datatype, String language) {
+    System.out.println("default_graph(...)");
```
lewismc: Using ```System``` calls is never particularly safe... are all of these calls necessary?

librdfa-rdf4j/src/main/c/rdfa.i
```diff
@@ -0,0 +1,40 @@
+%module(directors="1") rdfa
```
lewismc: Missing license header.

librdfa-rdf4j/src/main/resources/META-INF/services/org.eclipse.rdf4j.rio.RDFParserFactory
```diff
@@ -0,0 +1 @@
+org.apache.any23.rdf.rdfa.LibrdfaRDFaParserFactory
```
lewismc: License header if possible.

test-resources/src/test/resources/html/rdfa/basic.html
```diff
@@ -14,7 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<html xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
```
lewismc: Seems like a strange addition. Can you explain?

JulioCCBUcuenca:
Thanks @lewismc for the feedback. I will work on all your suggestions.

librdfa-rdf4j/README.MD
JulioCCBUcuenca: Each metric is in milliseconds.

librdfa-rdf4j/src/main/c/RdfaParser.h
JulioCCBUcuenca: The constructor is between lines 68-71. C++ uses a destructor function when the lifetime of an object ends. Here, the context stored in memory and the callback function are freed from memory.

librdfa-rdf4j/src/main/c/main.java
JulioCCBUcuenca: I am sorry, this file was intended for testing purposes when I was testing the bridge interaction. I will deleted.

test-resources/src/test/resources/html/rdfa/basic.html
JulioCCBUcuenca: The` lang` tag is used in HTML files, and the language of XHTML files is provided using `xml:lang`. So,  since this is a HTML page, it needs to use the lang tag. Current XML parsers can use both lang or xml:lang, but since librdfa uses an old library for parsing XML it generates an error since it cannot identify the language.

core/pom.xml
JulioCCBUcuenca: Since we haven't published the librdfa-rdf4j parser in maven central, I added the jars there. I added that because if while compiling librdfa is not installed, the build will fail. In that URL lives the jars of librdfa-rdf4j.

core/pom.xml
JulioCCBUcuenca: So, should I include librdfa-rdf4j as part of the build life-cycle?

core/pom.xml
JulioCCBUcuenca: Currently, librdfa-rdf4j is a standalone module that is not part of any23. Thus, at compiling time, any23 doesn't know about librdfa-rdf4j. If we integrate this with any, we need to deal with the requirement that librdfa-rdfa needs librdfa.

lewismc: librdfa-rdf4j should definitely be a separate module of Any23, please make sure that it is and that the require JAR dependencies are located at build and compile/runtime such.

test-resources/src/test/resources/html/rdfa/basic.html
lewismc: Thats fine thanks for the explanation.

lewismc:
@JulioCCBUcuenca Please address the issue of making librdfa-rdf4j a module within Any23 then push your changes. At that stage I will most likely approve the PR and we can test more thoroughly. Thanks.

HansBrende:
`Current XML parsers can use both lang or xml:lang, but since librdfa uses an old library for parsing XML it generates an error since it cannot identify the language.`

That seems worrisome to me... considering then, that most html pages will break the librdfa parser. 

@lewismc should we not test more thoroughly *before* merging to master? Maybe a separate branch instead?

Also, I would think that all of our current rdfa parsing tests should be duplicated for the librdfa parser to ensure that it is *at least as stable as our current rdfa parser*.  TBH, I'd be in favor of adding this into version 2.4 rather than 2.3 so we have more time to thoroughly test the module.

lewismc:
> @lewismc should we not test more thoroughly before merging to master? Maybe a separate branch instead?

Yes absolutely. We should also definitely extend the integration tests as well such that we can make more confident comparisons regarding runtime execution.

lewismc:
> TBH, I'd be in favor of adding this into version 2.4 rather than 2.3 so we have more time to thoroughly test the module.

That is fine with me...

JulioCCBUcuenca:
Thanks for your comments @lewismc. Will do it! :smile:  

@HansBrende librdfa supports RDFa 1.0 and 1.1, so I made sure that librdfa extractor works on all tests that Semargl extractor runs for both RDFa 1.0 and 1.1; additionally, the librdfa extractor is tested on the base suitcase (`AbstractRDFaExtractorTestCase` ). 

I agree that librdfa functionalities should go to other release. I am mostly concerned for the requirement that librdfa-rdf4j needs, which is that librdfa (the C library) needs to be installed. Remember, that not only the libraries that librdfa uses are old but the librdfa project itself has not being active in a while.

If we are moving to other branch, could someone create a new branch, so I can point the PR to the new branch.

lewismc:
> Remember, that not only the libraries that librdfa uses are old but the librdfa project itself has not being active in a while.

Yes but it is compliant with the RDFa test suite which is exactly what we want. It can be assumed to be somewhat complete...

> If we are moving to other branch, could someone create a new branch, so I can point the PR to the new branch.

@HansBrende can you please create a new branch called ANY23-295, I do not have access to a terminal right now. Thank you in advance, 

> I will be glad to keep contributing after GSoC. Honestly, I'd like the idea of becoming a commiter, so I have to keep working to earn it.

👍

HansBrende:
@JulioCCBUcuenca Great to hear it passes the semargl test suites! However, `AbstractRDFaExtractorTestCase` contains only a fraction of the tests we run on RDFa. You should also test against the [RDFaExtractorTest](https://github.com/apache/any23/blob/master/core/src/test/java/org/apache/any23/extractor/rdfa/RDFaExtractorTest.java) and [RDFa11ExtractorTest](https://github.com/apache/any23/blob/master/core/src/test/java/org/apache/any23/extractor/rdfa/RDFa11ExtractorTest.java) classes.

JulioCCBUcuenca:
@HansBrende it is testing those main tests as well. It is testing the tests of Semargl RDFa 1.0 (`RDFaExtractorTest`), Semargl RDFa 1.1 (`RDFa11ExtractorTest`), and base suitcase `AbstractRDFaExtractorTestCase`.

HansBrende:
@JulioCCBUcuenca Alright, sounds great then!

HansBrende:
@lewismc I attempted to create a new branch using:

```
git checkout -b ANY23-295
git push canonical ANY23-295
```
to which git responded with:
```
To https://git-wip-us.apache.org/repos/asf/any23.git
 * [new branch]        ANY23-295 -> ANY23-295
```
However, it isn't showing up under "Branches". Not sure why. It's showing up under my own "Branches" when I pushed to `origin`.

HansBrende:
Ok, apparently due to some quirkiness of the way mirroring works, new branches in git-wip will not show up on github until an actual change is made to the branch.

HansBrende:
Ok, I added some whitespace to a package-info.java file, should be showing up now.

JulioCCBUcuenca:
I change the base branch. Thanks so much @HansBrende

librdfa-rdf4j/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  
+  <parent>
+    <groupId>org.apache.any23</groupId>
+    <artifactId>apache-any23</artifactId>
+    <version>2.3-SNAPSHOT</version>
```
lewismc: @JulioCCBUcuenca can you please update this to 2.4-SNAPSHOT so it is synchronized with master branch. Thank you

HansBrende:
@lewismc do we have any benchmarks on this version vs our current rdfa parser?

lewismc:
@HansBrende

https://cwiki.apache.org/confluence/display/ANY23/Librdfa-rdf4j+documentation#Librdfa-rdf4jdocumentation-Performance 

The current implementation does not look particularly favorable for librdfa

HansBrende:
@lewismc I've resolved that issue, moving the semargl-specific bugfixes into the semargl extractors. 

@JulioCCBUcuenca a couple recommendations for this branch whenever you get a chance:

(1) Synchronize this branch with master
(2) Rerun test suite, making sure all tests still pass
(3) If possible, benchmark the `Extractor` wrappers (not just underlying rdf4j parsers). Doing this may give the librdfa extractor a performance edge, as the semargl parser requires the html stream to be preprocessed and transformed into strictly conforming XML.

lewismc:
@HansBrende 

> My first thought is: if the performance of this module is not as good as that of our current implementation, then in its current form, what is the added value?

The project was started and completed as a research effort to see what results we could come up with. Thank you so much for the additional investigation on your part... I agree with your statement. I'll see if I can find some time to update the PR. Thanks

PR #116, opened 2018-08-16
Any23 388

larsgsvensson:
Making the constructor of `RDFWriterTripleHandler` and its `writer` field protected, as [discussed](https://issues.apache.org/jira/browse/ANY23-388?focusedCommentId=16580679&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16580679) by me, @lewismc and @HansBrende


HansBrende:
@larsgsvensson Thanks for the pull request! It looks like some of your commits reverse recent changes made to the master branch. You might want to just start over with a clean pull from master.

lewismc:
@HansBrende feel free to merge once you are satisfied.

core/src/main/java/org/apache/any23/writer/RDFWriterTripleHandler.java
```diff
@@ -35,7 +35,7 @@
  */
 public abstract class RDFWriterTripleHandler implements FormatWriter, TripleHandler {
 
-    private final RDFWriter writer;
+    protected final RDFWriter writer;
```
jgrzebyta: I am a bit confused why you just did not add setter instead having the filed protected. Or expolit the protected constructor you did below.

core/src/main/java/org/apache/any23/writer/RDFWriterTripleHandler.java
```diff
@@ -44,7 +44,7 @@
      */
     private boolean annotated = false;
 
-    RDFWriterTripleHandler(RDFWriter destination) {
+    protected RDFWriterTripleHandler(RDFWriter destination) {
```
jgrzebyta: I like it.

core/src/main/java/org/apache/any23/writer/RDFWriterTripleHandler.java
HansBrende: @jgrzebyta I don't see any reason to disallow protected access to the `writer` if we allow protected access to the constructor. Subclasses could bypass the `private` modifier anyway by simply saving a reference to the writer in the constructor.

larsgsvensson:
@HansBrende scripsit:
> Thanks for the pull request! It looks like some of your commits reverse recent changes made to the master branch. You might want to just start over with a clean pull from master.

Hmm, funny. I rebased just before creating the PR... 

As I'm still somewhat of a GitHub newbie: Does that mean that I should delete the branch (and then what happens to this PR)?

core/src/main/java/org/apache/any23/writer/RDFWriterTripleHandler.java
jgrzebyta: @HansBrende There is no reason. It is only the coding style.

But if user will override constructor in her/his custom class will call `super(someWriter)` anyway. In that case constructor is a setter for the writer.

HansBrende:
@larsgsvensson it looks like you may have rebased in the wrong direction? One problem is that your master branch is 3 commits ahead of the apache/master branch. So here's what I would do (before making any more commits):

``` bash
git checkout master
git reset --hard HEAD~3
git pull https://github.com/apache/any23.git master
git push -f origin master

git checkout ANY23-388
git reset --hard HEAD~5
git rebase master
git push -f origin ANY23-388
```

After you do that, then make your changes to the ANY23-388 branch. Then:

```bash
git add .
git commit -m "ANY23-388 [message]"
git push origin ANY23-388
```

Then we should be good to go.

larsgsvensson:
@HansBrende Thanks, now it should be OK. New PR #117 submitted

PR #121, opened 2018-09-06
ANY23-396 Add ability to run extractors in flow 

jgrzebyta:
- Add unit test with example. The test contains expected model.

### DIFF
--- a/api/src/main/resources/default-configuration.properties
+++ b/api/src/main/resources/default-configuration.properties
@@ -76,3 +76,6 @@ any23.extraction.csv.comment=#
 # A confidence threshold for the OpenIE extractions
 # Any extractions below this value will not be processed.
 any23.extraction.openie.confidence.threshold=0.5
+
+# Allows to enable(on)/disable(off) the workflow feature
+any23.extraction.workflows=off
--- a/cli/src/main/java/org/apache/any23/cli/Rover.java
+++ b/cli/src/main/java/org/apache/any23/cli/Rover.java
@@ -137,10 +137,6 @@ public class Rover extends BaseTool {
             );
         }
 
-        if (workflow) {
-            tripleHandler = new BufferedTripleHandler(tripleHandler);
-        }
-
         if (logFile != null) {
             try {
                 tripleHandler = new LoggingTripleHandler(tripleHandler, new PrintWriter(logFile));
@@ -158,6 +154,10 @@ public class Rover extends BaseTool {
             tripleHandler = new IgnoreAccidentalRDFa(new IgnoreTitlesOfEmptyDocuments(tripleHandler),
                                                      true    // suppress stylesheet triples.
                                                      );
+        }
+
+        if (workflow) {
+            tripleHandler = new BufferedTripleHandler(tripleHandler);
         }
 
         reportingTripleHandler = new ReportingTripleHandler(tripleHandler);
--- a/cli/src/main/java/org/apache/any23/cli/Rover.java
+++ b/cli/src/main/java/org/apache/any23/cli/Rover.java
@@ -100,8 +100,8 @@ public class Rover extends BaseTool {
     @Parameter(names = { "-d", "--defaultns" }, description = "Override the default namespace used to produce statements.")
     private String defaultns;
 
-    @Parameter(names = "--workflows", description = "Run extractors in workflow.")
-    private boolean workflows = false;
+    @Parameter(names = "--workflow", description = "Run extractors in workflow.")
+    private boolean workflow = false;
 
     // non parameters
 
@@ -137,7 +137,7 @@ public class Rover extends BaseTool {
             );
         }
 
-        if (workflows) {
+        if (workflow) {
             tripleHandler = new BufferedTripleHandler(tripleHandler);
         }
 
@@ -173,6 +173,8 @@ public class Rover extends BaseTool {
             extractionParameters.setProperty(ExtractionParameters.EXTRACTION_CONTEXT_IRI_PROPERTY,
                                              defaultns);
         }
+
+        extractionParameters.setFlag(ExtractionParameters.EXTRACTION_WORKFLOWS_FLAG, workflow);
 
         any23 = (extractors.isEmpty()) ? new Any23()
                                                    : new Any23(extractors.toArray(new String[extractors.size()]));
--- a/cli/src/main/java/org/apache/any23/cli/Rover.java
+++ b/cli/src/main/java/org/apache/any23/cli/Rover.java
@@ -30,12 +30,7 @@ import org.apache.any23.extractor.ExtractionParameters.ValidationMode;
 import org.apache.any23.filter.IgnoreAccidentalRDFa;
 import org.apache.any23.filter.IgnoreTitlesOfEmptyDocuments;
 import org.apache.any23.source.DocumentSource;
-import org.apache.any23.writer.BenchmarkTripleHandler;
-import org.apache.any23.writer.LoggingTripleHandler;
-import org.apache.any23.writer.ReportingTripleHandler;
-import org.apache.any23.writer.TripleHandler;
-import org.apache.any23.writer.TripleHandlerException;
-import org.apache.any23.writer.WriterFactoryRegistry;
+import org.apache.any23.writer.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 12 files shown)


jgrzebyta:
I guess I have finished the first version. Any comments?

core/src/main/java/org/apache/any23/writer/BufferedTripleHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public BufferedTripleHandler(TripleHandler underlying) {
+        this.underlying = underlying;
+
+        // hide model in the thread
+        WorkflowContext wc = new WorkflowContext(underlying);
+        BufferedTripleHandler.workflowContext.set(wc);
+    }
+
+    private static final ThreadLocal<WorkflowContext> workflowContext = new ThreadLocal<>();
```
HansBrende: Model should not be static, unless there is a very good reason for doing so?

core/src/main/java/org/apache/any23/extractor/SingleDocumentExtraction.java
```diff
@@ -295,6 +294,12 @@ public SingleDocumentExtractionReport run(ExtractionParameters extractionParamet
         } finally {
 	        try {
 	            output.endDocument(documentIRI);
+
+	            // in case of workflow flag release data from model
+                if (extractionParameters.getFlag(ExtractionParameters.EXTRACTION_WORKFLOWS_FLAG)) {
+                    BufferedTripleHandler.releaseModel();
```
HansBrende: This should not be a static call.

core/src/main/java/org/apache/any23/extractor/SingleDocumentExtraction.java
```diff
@@ -483,6 +488,14 @@ private SingleExtractionReport runExtractor(
                         documentReport.getDocument(),
                         extractionResult
                 );
+            } else if (extractor instanceof ModelExtractor) {
+                final ModelExtractor modelExtractor = (ModelExtractor) extractor;
+                final Model singleModel = BufferedTripleHandler.getModel();
```
HansBrende: Should not be static.

core/src/main/java/org/apache/any23/writer/BufferedTripleHandler.java
```diff
@@ -0,0 +1,161 @@
...
+     * Releases content of the model into underlying writer.
+     */
+    public static void releaseModel() throws TripleHandlerException {
+        if(!BufferedTripleHandler.isDocumentFinish) {
+            throw new RuntimeException("Before releasing document should be finished.");
+        }
+
+        WorkflowContext workflowContext = BufferedTripleHandler.workflowContext.get();
+
+        String lastExtractor = ((Stack<String>) workflowContext.extractors).peek();
```
HansBrende: Feels hacky... what if not all of the triples came from the same extractor?

cli/src/main/java/org/apache/any23/cli/Rover.java
```diff
@@ -172,6 +174,8 @@ protected void configure() {
                                              defaultns);
         }
 
+        extractionParameters.setFlag(ExtractionParameters.EXTRACTION_WORKFLOWS_FLAG, workflow);
```
HansBrende: We should not need a separate flag to enable certain extractors. If an extractor is contained within the extractor group we are using, then that should be, on its own, enough to enable itself.

api/src/main/resources/default-configuration.properties
```diff
@@ -76,3 +76,6 @@ any23.extraction.csv.comment=#
 # A confidence threshold for the OpenIE extractions
 # Any extractions below this value will not be processed.
 any23.extraction.openie.confidence.threshold=0.5
+
+# Allows to enable(on)/disable(off) the workflow feature
+any23.extraction.workflows=off
```
HansBrende: No extra flag should be needed for this.

HansBrende:
Aside from the comments I've made on this PR, I'm still not convinced that having a ModelExtractor is a good idea in the first place. Why not just create a ModelWriter (as in ANY23-397) or an equivalent "collecting" TripleHandler, and then allow the end user to transform the collected statements however they wish?

Having a ModelExtractor creates additional questions & complexities: in what order are the extractors executed? (Certainly the ModelExtractors would have to be executed last in order to have access to all previously collected statements.) What if multiple ModelExtractors are declared? Which ones have higher precedence in the extraction order?

I'm not sure that having a dedicated ModelExtractor is worth the trouble of dealing with these complexities, when a user could accomplish the same thing by simply transforming the statements collected by a ModelWriter or equivalent, or defining their own filtering and/or mapping TripleHandler.

HansBrende:
Another thought:

Using the TripleHandler interface (as intended) to transform triples, rather than a separate ModelExtractor, has the added advantage that the triples might not necessarily need to be stored in memory during the transformation process. The user could implement either a "collecting" triple handler which stores statements in memory prior to transforming them, or a "streaming" triple handler for transformation-on-the-fly (e.g., if mapping some predicate A to some other predicate B), or some combination of these two concepts. The "collecting" ability could be easily supplemented with a `ModelWriter` or equivalent, as in [ANY23-397](https://issues.apache.org/jira/browse/ANY23-397).

But adding a separate "ModelExtractor" concept only muddles this already-existing ability to transform triples with TripleHandlers by introducing a redundant construct of more limited abstraction power than what already exists.

So for me:
-1 for ANY23-396
+1 for ANY23-397

@lewismc any thoughts?

core/src/main/java/org/apache/any23/writer/BufferedTripleHandler.java
jgrzebyta: Yes I agree with you. The idea is that these models should be presented later on (inside SingleDocumentWriter) to ModelExtractor and conrain parsing outcome of previous extractors. Access to those models is not propagated further down from Rover. I meant without changing api. I thought adding them into extraction parameters.

core/src/main/java/org/apache/any23/writer/BufferedTripleHandler.java
jgrzebyta: The idea is to find extractor which produces theustomer's domain-specific rdf graph.

HansBrende: @jgrzebyta IMHO, it would be vastly more straightforward to simply have the user extend the [`CompositeTripleHandler`](https://github.com/apache/any23/blob/master/core/src/main/java/org/apache/any23/writer/CompositeTripleHandler.java) class to filter and transform triples into a domain-specific rdf graph of their choosing, before delegating the final domain-specific triple outputs to the wrapped `TripleHandler` instance(s) by calling `super.receiveTriple( [modified triple] )`.

(Analogous in concept to Java's own [`FilterOutputStream`](https://docs.oracle.com/javase/8/docs/api/java/io/FilterOutputStream.html) class.)

HansBrende:
@jgrzebyta But as far as rover goes, you're right: we currently don't have support for using an arbitrary triple handler. Looks like it expects an RDFFormat and then finds a triple handler based on that. 

I wonder if it would be possible to allow a more flexible way to specify triple handlers as rover arguments to fix this problem? While I don't think that creating a `ModelExtractor` as currently defined in this PR is the way to go, I do think that rover needs to be improved in this respect.

I will think on this.

HansBrende:
@jgrzebyta One easy solution to my above comment that I can think of right off the bat is as follows:

First, we could extend the WriterFactory interface as follows (or similar):

```
interface DelegatingWriterFactory extends WriterFactory {
    TripleHandler getWriter(TripleHandler delegate);
}
```

Second, in the rover `--format` flag (which actually accepts a WriterFactory *id*, not necessarily a format name), we could simply allow a comma-separated *list* of WriterFactory ids rather than a single id. Then, to construct the final writer, we'd compose each writer in the list with the previous one, i.e.:

```
Collections.reverse(listOfIds);

tripleHandler = getWriterFactoryForId(listOfIds.get(0)).getRdfWriter(outputStream);

for (String id : listOfIds.subList(1, listOfIds.size())) {
    tripleHandler = ((DelegatingWriterFactory)getWriterFactoryForId(id)).getWriter(tripleHandler);
}
```
This is just one initial idea, but food for thought. It also seems more in line with your concept of a "flow".

What do you think?

HansBrende:
@jgrzebyta please check out [this PR](https://github.com/apache/any23/pull/122), which is an implementation of what I've just described.

It seems like a lot simpler and less error-prone way to produce a domain-specific rdf graph.

Eager to know your thoughts!

HansBrende:
Now that ANY23-396 has been implemented in #122 and merged into master, can we close this PR? @lewismc ? @jgrzebyta ? I don't have the required permissions to close issues myself.

PR #133, opened 2019-01-30
ANY23-423: Update POM on the move to gitbox

afs:


lewismc:
+1 please merge here @afs thank you

PR #134, opened 2019-01-31
ANY23-422 Error message when any23 cli tool used

davidcockbill:
This PR fixes the issue in question. However I am unaware of whether this will have any unwanted side-effects.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -347,7 +347,6 @@
             </program>
           </programs>
           <configurationDirectory>conf</configurationDirectory>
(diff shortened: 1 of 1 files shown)


HansBrende:
Looks good to me, although to be honest, I am unfamiliar with this plugin and why we were setting the `configurationSourceDirectory` to the `src/test/resources` folder (I think the default value for that setting is `src/main/config`). 

Thank you for your work on narrowing down this bug!

@ansell do you have any comments on this PR before we merge it?

PR #143, opened 2019-09-24
ANY23-183 Address javac warning's in Any23 code base

lewismc:
This issue addresses https://issues.apache.org/jira/browse/ANY23-183

You can also see I bumped the javac version to 13... which is what I am using across the board now. If this is not desired then i can roll this back.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,8 +263,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <javac.src.version>13</javac.src.version>
-    <javac.target.version>13</javac.target.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -263,8 +263,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <javac.src.version>1.8</javac.src.version>
-    <javac.target.version>1.8</javac.target.version>
+    <javac.src.version>13</javac.src.version>
+    <javac.target.version>13</javac.target.version>
```
HansBrende: Can we keep this at 1.8? I don't think we should increase the java version until we are forced to do so by one of our dependencies. There are a *lot* of applications still on java 8.

lewismc: I agree...

PR #144, opened 2019-09-25
ANY23-433 upgrade rdf4j to v3.0.0

HansBrende:
Upgrade rdf4j to v3.0.0 and remove jsonld hack. I also took out the rdf4j middleman when working with the jsonld-java library, which will allow us to more easily tinker with jsonld-java in the future instead of resorting to terrible hacks.

### DIFF
--- a/core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
@@ -96,7 +96,7 @@ public abstract class BaseRDFExtractor implements Extractor.ContentExtractor {
         }
     }
 
-    // keep private to avoid backwards compatibility woes (may move around later)
+    // keep package-access to avoid backwards compatibility woes if protected (may move around later)
     @SuppressWarnings("Duplicates")
     static String toString(Throwable th) {
         StringWriter writer = new StringWriter();
--- a/core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
+++ b/core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
@@ -17,15 +17,12 @@
 
 package org.apache.any23.extractor.rdf;
 
-import com.fasterxml.jackson.core.JsonLocation;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import org.apache.any23.extractor.ExtractionContext;
 import org.apache.any23.extractor.ExtractionException;
 import org.apache.any23.extractor.ExtractionParameters;
 import org.apache.any23.extractor.ExtractionResult;
 import org.apache.any23.extractor.Extractor;
 import org.apache.any23.extractor.IssueReport;
-import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.RDFParser;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


HansBrende:
@lewismc any comments?

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
```diff
@@ -93,36 +90,15 @@ public void run(
...
-                    extractionResult.notifyIssue(IssueReport.IssueLevel.FATAL, err.getOriginalMessage(), loc.getLineNr(), loc.getColumnNr());
-                }
-            } else {
-                extractionResult.notifyIssue(IssueReport.IssueLevel.FATAL, toString(ex), -1, -1);
-            }
+            extractionResult.notifyIssue(IssueReport.IssueLevel.FATAL, toString(ex), -1, -1);
         }
     }
 
     // keep private to avoid backwards compatibility woes (may move around later)
```
lewismc: This comment should be edited or removed

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
```diff
@@ -93,36 +90,15 @@ public void run(
...
-
-            RDFFormat format = parser.getRDFFormat();
-
-            if (format.hasFileExtension("jsonld") || format.hasMIMEType("application/ld+json")) {
-                in = new JsonCleaningInputStream(in);
-            }
-
             parser.parse(in, extractionContext.getDocumentIRI().stringValue());
         } catch (Exception ex) {
-            // ANY23-420: jsonld-java can sometimes throw IllegalArgumentException,
```
lewismc: I like your retro-active move of this code down into ```JSONLDExtractor.java```

core/src/main/java/org/apache/any23/extractor/rdf/JSONLDJavaSink.java
```diff
@@ -0,0 +1,114 @@
...
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.ValueFactory;
+import com.github.jsonldjava.core.JsonLdConsts;
+import com.github.jsonldjava.core.JsonLdTripleCallback;
+import com.github.jsonldjava.core.RDFDataset;
+
+/**
+ * @author Hans Brende (hansbrende@apache.org)
+ */
+class JSONLDJavaSink implements JsonLdTripleCallback {
```
lewismc: I like it. I don't know quite enough about jsonld-java so know that this existed. Excellent suggestion.

core/src/main/java/org/apache/any23/extractor/rdf/BaseRDFExtractor.java
```diff
@@ -93,36 +90,15 @@ public void run(
...
-                }
-            } else {
-                extractionResult.notifyIssue(IssueReport.IssueLevel.FATAL, toString(ex), -1, -1);
-            }
+            extractionResult.notifyIssue(IssueReport.IssueLevel.FATAL, toString(ex), -1, -1);
         }
     }
 
     // keep private to avoid backwards compatibility woes (may move around later)
     @SuppressWarnings("Duplicates")
```
lewismc: Can also remove this ```Duplicates``` annotation it is now unnecessary

HansBrende: Actually we still need that annotation to avoid Intellij warnings... unfortunately the `BaseRDFaExtractor` still can't see it because it's in a different package, and I don't want to make it protected, otherwise it'll be a backwards incompatible change if we delete it later (which we will probably do).

PR #146, opened 2019-09-26
ANY23-426 Address Javadoc WARNING's 

lewismc:
This PR addresses https://issues.apache.org/jira/browse/ANY23-426

### DIFF
--- a/api/src/main/java/org/apache/any23/configuration/Setting.java
+++ b/api/src/main/java/org/apache/any23/configuration/Setting.java
@@ -113,6 +113,8 @@ public abstract class Setting<V> implements Cloneable {
     }
 
     /**
+     * @param setting a setting key to retrieve
+     * @param <S> generic type for setting
      * @return this setting, if this setting has the same key as the supplied setting
      */
     @SuppressWarnings("unchecked")
@@ -121,6 +123,7 @@ public abstract class Setting<V> implements Cloneable {
     }
 
     /**
+     * @param newValue a value to provide for the new setting pair
      * @return a new {@link Setting} object with this setting's key and the supplied value.
      *
      * @throws IllegalArgumentException if the new value was invalid, as determined by:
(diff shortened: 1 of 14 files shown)


api/src/main/java/org/apache/any23/configuration/Setting.java
```diff
@@ -113,6 +113,8 @@ public final Type getValueType() {
     }
 
     /**
+     * @param setting a setting key to retrieve
+     * @param <S> generic type for setting
```
HansBrende: Perhaps something more along the lines of:

```
@param setting a setting that may or may not have the same key as this setting
@param <S> the type of the supplied setting
```

api/src/main/java/org/apache/any23/configuration/Setting.java
```diff
@@ -121,6 +123,7 @@ public final Type getValueType() {
     }
 
     /**
+     * @param newValue a value to provide for the new setting pair
```
HansBrende: Simply 
```
a value for a new setting
```
should suffice.

api/src/main/java/org/apache/any23/configuration/Settings.java
```diff
@@ -132,7 +137,8 @@ public static Settings of(Setting<?>... settings) {
     }
 
     /**
-     * Returns a {@link Settings} object containing the supplied settings.
+     * @param c a polulated {@link java.util.Collection} of {@link org.apache.any23.configuration.Setting}'s
```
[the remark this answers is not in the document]
lewismc: Thanks

api/src/main/java/org/apache/any23/configuration/Setting.java
lewismc: Thanks

api/src/main/java/org/apache/any23/configuration/Setting.java
lewismc: Thanks

lewismc:
Thanks for review @HansBrende

PR #151, opened 2019-10-16
ANY23-445 Review spotbugs issues

lewismc:
First pass at addressing https://issues.apache.org/jira/browse/ANY23-445.
Still loads to do.

### DIFF
--- a/api/src/main/java/org/apache/any23/vocab/HCard.java
+++ b/api/src/main/java/org/apache/any23/vocab/HCard.java
@@ -36,10 +36,9 @@ public class HCard extends Vocabulary {
         return instance;
     }
 
-    public IRI Card  = createClass(NS, "Card");
-    public IRI Address   = createClass(NS, "Address");
-    public IRI Geo = createClass(NS, "Geo");
-
+    public IRI card  = createClass(NS, "Card");
+    public IRI address   = createClass(NS, "Address");
+    public IRI _Geo = createClass(NS, "Geo");
 
     public IRI name  = createProperty(NS, "name");
     public IRI honorific_prefix   = createProperty(NS, "honorific-prefix");
--- a/api/src/main/java/org/apache/any23/vocab/HEntry.java
+++ b/api/src/main/java/org/apache/any23/vocab/HEntry.java
@@ -37,7 +37,7 @@ public class HEntry extends Vocabulary {
         return instance;
     }
 
-    public IRI Entry  = createClass(NS, "Entry");
+    public IRI entry  = createClass(NS, "Entry");
     public IRI author   = createClass(NS, "author");
     public IRI location = createClass(NS, "location");
 
@@ -51,7 +51,7 @@ public class HEntry extends Vocabulary {
     public IRI url   = createProperty(NS, "url");
     public IRI uid  = createProperty(NS, "uid");
     public IRI syndication   = createProperty(NS, "syndication");
-    public IRI in_reply_to   = createProperty(NS, "in-reply-to");
+    public IRI inReplyTo   = createProperty(NS, "in-reply-to");
 
     private HEntry() {
         super(NS);
--- a/api/src/main/java/org/apache/any23/vocab/HItem.java
+++ b/api/src/main/java/org/apache/any23/vocab/HItem.java
@@ -37,7 +37,7 @@ public class HItem extends Vocabulary {
         return instance;
     }
 
-    public IRI Item  = createClass(NS, "Item");
+    public IRI item  = createClass(NS, "Item");
     public IRI name  = createProperty(NS, "name");
     public IRI url   = createProperty(NS, "url");
     public IRI photo = createProperty(NS, "photo");
--- a/api/src/main/java/org/apache/any23/vocab/HRecipe.java
+++ b/api/src/main/java/org/apache/any23/vocab/HRecipe.java
@@ -38,10 +38,10 @@ public class HRecipe extends Vocabulary {
     }
 
     // Resources.
-    public IRI Recipe     = createClass(NS, "Recipe");
-    public IRI Duration   = createClass(NS, "Duration");
-    public IRI Ingredient = createClass(NS, "Ingredient");
-    public IRI Nutrition  = createClass(NS, "Nutrition");
+    public IRI recipe     = createClass(NS, "Recipe");
+    public IRI _Duration   = createClass(NS, "Duration");
+    public IRI _Ingredient = createClass(NS, "Ingredient");
+    public IRI _Nutrition  = createClass(NS, "Nutrition");
 
     // Properties.
     public IRI fn                     = createProperty(NS, "fn");
--- a/api/src/main/java/org/apache/any23/vocab/HResume.java
+++ b/api/src/main/java/org/apache/any23/vocab/HResume.java
@@ -36,12 +36,11 @@ public class HResume extends Vocabulary {
         return instance;
     }
 
-    public IRI Resume  = createClass(NS, "Resume");
+    public IRI resume  = createClass(NS, "Resume");
     public IRI education   = createClass(NS, "education");
     public IRI experience = createClass(NS, "experience");
     public IRI contact = createClass(NS, "contact");
     public IRI affiliation = createClass(NS, "affiliation");
-
 
     public IRI name  = createProperty(NS, "name");
     public IRI summary   = createProperty(NS, "summary");
--- a/api/src/main/java/org/apache/any23/vocab/Vocabulary.java
+++ b/api/src/main/java/org/apache/any23/vocab/Vocabulary.java
@@ -142,12 +142,12 @@ public abstract class Vocabulary {
      */
     public IRI getPropertyCamelCase(String property) {
         String[] names = property.split("\\W");
-        String camelCase = names[0];
+        StringBuffer camelCase = new StringBuffer().append(names[0]);
         for (int i = 1; i < names.length; i++) {
             String tmp = names[i];
-            camelCase += tmp.replaceFirst("(.)", tmp.substring(0, 1).toUpperCase(java.util.Locale.ROOT));
+            camelCase.append(tmp.replaceFirst("(.)", tmp.substring(0, 1).toUpperCase(java.util.Locale.ROOT)));
         }
-        return getProperty(camelCase);
+        return getProperty(camelCase.toString());
     }
 
     /**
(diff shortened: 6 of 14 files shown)


api/src/main/java/org/apache/any23/vocab/Vocabulary.java
```diff
@@ -142,12 +142,12 @@ public IRI getProperty(String name, IRI defaultValue) {
      */
     public IRI getPropertyCamelCase(String property) {
         String[] names = property.split("\\W");
-        String camelCase = names[0];
+        StringBuffer camelCase = new StringBuffer().append(names[0]);
```
HansBrende: `StringBuilder` would be a better choice here; it doesn't need synchronization.

`StringBuilder camelCase = new StringBuilder(property.length()).append(names[0]);

api/src/main/java/org/apache/any23/vocab/HResume.java
```diff
@@ -36,13 +36,12 @@ public static HResume getInstance() {
         return instance;
     }
 
-    public IRI Resume  = createClass(NS, "Resume");
+    public IRI resume  = createClass(NS, "Resume");
```
HansBrende: This is a breaking change. Probably requires more discussion.

api/src/main/java/org/apache/any23/vocab/HRecipe.java
```diff
@@ -38,10 +38,10 @@ public static HRecipe getInstance() {
...
 
     // Resources.
-    public IRI Recipe     = createClass(NS, "Recipe");
-    public IRI Duration   = createClass(NS, "Duration");
-    public IRI Ingredient = createClass(NS, "Ingredient");
-    public IRI Nutrition  = createClass(NS, "Nutrition");
+    public IRI recipe     = createClass(NS, "Recipe");
+    public IRI _Duration   = createClass(NS, "Duration");
+    public IRI _Ingredient = createClass(NS, "Ingredient");
+    public IRI _Nutrition  = createClass(NS, "Nutrition");
```
HansBrende: Breaking changes. Probably require more discussion.

api/src/main/java/org/apache/any23/vocab/HItem.java
```diff
@@ -37,7 +37,7 @@ public static HItem getInstance() {
         return instance;
     }
 
-    public IRI Item  = createClass(NS, "Item");
+    public IRI item  = createClass(NS, "Item");
```
HansBrende: Breaking change. Probably requires some discussion.

api/src/main/java/org/apache/any23/vocab/HEntry.java
```diff
@@ -51,7 +51,7 @@ public static HEntry getInstance() {
     public IRI url   = createProperty(NS, "url");
     public IRI uid  = createProperty(NS, "uid");
     public IRI syndication   = createProperty(NS, "syndication");
-    public IRI in_reply_to   = createProperty(NS, "in-reply-to");
+    public IRI inReplyTo   = createProperty(NS, "in-reply-to");
```
HansBrende: Breaking change. Probably requires discussion.

api/src/main/java/org/apache/any23/vocab/HEntry.java
```diff
@@ -37,7 +37,7 @@ public static HEntry getInstance() {
         return instance;
     }
 
-    public IRI Entry  = createClass(NS, "Entry");
+    public IRI entry  = createClass(NS, "Entry");
```
HansBrende: Breaking change. Probably requires discussion.

api/src/main/java/org/apache/any23/vocab/HCard.java
```diff
@@ -36,10 +36,9 @@ public static HCard getInstance() {
         return instance;
     }
 
-    public IRI Card  = createClass(NS, "Card");
-    public IRI Address   = createClass(NS, "Address");
-    public IRI Geo = createClass(NS, "Geo");
-
+    public IRI card  = createClass(NS, "Card");
+    public IRI address   = createClass(NS, "Address");
+    public IRI _Geo = createClass(NS, "Geo");
```
HansBrende: These are breaking changes. We probably don't want to release another major version for this, so if we want to change the naming, we should probably deprecate the old ones rather than removing them.

But in any case, how is `_Geo` an improvement over `Geo`? Not sure I agree with that naming change...

HansBrende: By "breaking change", simply I mean a backwards-incompatible change (see https://semver.org/ for more info).

I would say, it's probably a good idea to disable the capitalization check for now, since so many parts of our codebase are already written the way they are.

lewismc: Reverted this and the rest and disabled check. I agree with you.

lewismc:
Current branch results in following errors
```
[INFO] --- spotbugs-maven-plugin:3.1.12.2:check (default-cli) @ apache-any23-api ---
[INFO] BugInstance size is 2
[INFO] Error size is 0
[INFO] Total bugs: 2
[ERROR] org.apache.any23.mime.MIMEType defines compareTo(MIMEType) and uses Object.equals() [org.apache.any23.mime.MIMEType] At MIMEType.java:[line 134] EQ_COMPARETO_USE_OBJECT_EQUALS
[ERROR] Possible null pointer dereference in org.apache.any23.plugin.Any23PluginManager.loadJARDir(File) due to return value of called method [org.apache.any23.plugin.Any23PluginManager, org.apache.any23.plugin.Any23PluginManager] Dereferenced at Any23PluginManager.java:[line 201]Known null at Any23PluginManager.java:[line 201] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE
```
@HansBrende if you get a chance can you review. You can run spotbugs analysis as follows
```
mvn spotbugs:check
```
You can also get detailed information by running
```
mvn clean install -DskipTests && mvn spotbugs:gui
```
You can then load the file in ./api/target/spotbugsXml.xml

api/src/main/java/org/apache/any23/vocab/Vocabulary.java
```diff
@@ -142,12 +142,12 @@ public IRI getProperty(String name, IRI defaultValue) {
      */
     public IRI getPropertyCamelCase(String property) {
         String[] names = property.split("\\W");
-        String camelCase = names[0];
+        StringBuilder camelCase = new StringBuilder(property.length()).append(names[0]);
         for (int i = 1; i < names.length; i++) {
             String tmp = names[i];
-            camelCase += tmp.replaceFirst("(.)", tmp.substring(0, 1).toUpperCase(java.util.Locale.ROOT));
+            camelCase.append(tmp.replaceFirst("(.)", tmp.substring(0, 1).toUpperCase(java.util.Locale.ROOT)));
```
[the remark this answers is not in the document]
lewismc: This would require that we import commons-lang3... is this something you are satisfied with considering that we were only just discussing dependency bloat. wdyt?

api/src/main/java/org/apache/any23/extractor/ExtractorGroup.java
```diff
@@ -78,6 +78,7 @@ public boolean allExtractorsSupportAllContentTypes() {
         return true;
     }
 
+    @SuppressWarnings("unlikely-arg-type")
```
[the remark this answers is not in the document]
lewismc: I've removed both the `SupressWarnings` javac annotation and thought about this one again.

I've decided to introduce `<omitVisitors>FindUnrelatedTypesInGenericContainer</omitVisitors>` which will omit spotbugs from identifying this issue as a bug. My justification is that, in this case, because the `ExtractorFactory#getSupportedMIMETypes()` supports wildcards, and as far as I can see there is currently no way to determine the full Collection (e.g. `*/*`) of supported `MimeType`'s then we cannot enforce another check whether an `ExtractorGroup` supports all `MimeType`'s. 

What do you think about this assessment?

api/src/main/java/org/apache/any23/vocab/Vocabulary.java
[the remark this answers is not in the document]
HansBrende: We already have a dependency on commons-lang3, and it is being used all over our codebase. So I'm not too worried about that dependency in particular.

api/src/main/java/org/apache/any23/extractor/ExtractorGroup.java
```diff
@@ -24,8 +24,8 @@
 import java.util.Iterator;
 
 /**
- * It simple models a group of {@link ExtractorFactory} providing
- * simple accessing methods.
+ * Models a group of {@link ExtractorFactory} objcts providing
```
HansBrende: typo

api/src/main/java/org/apache/any23/extractor/ExtractorGroup.java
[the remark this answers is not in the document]
HansBrende: Sorry, I wasn't clear. The code inside this method has a bug, as _correctly_ identified by the spotbugs plugin. The code is testing if a `Collection<MIMEType>` contains a `String`. This will always return false, as a `String` _cannot_ be an instance of `MIMEType`.

Therefore, the method needs to be fixed, and the spotbugs plugin is doing its job correctly: no suppression of warnings is needed here.

api/src/main/java/org/apache/any23/vocab/Vocabulary.java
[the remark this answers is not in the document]
HansBrende: I suppose it is true that this class in particular is part of any23 `api` rather than `core`, and `core` has the commons-lang3 dependency, so this would add that extra dependency to `api`. I'm not sure if this is a big problem since most people are using `core` rather than `api` alone... however if you'd rather not add an extra dependency to `api` I suppose we could leave it as-is for now (however I'd like to improve the implementation of that method at some point in the future).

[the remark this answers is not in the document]
HansBrende: And now that I think about it more, I agree that there is definitely something to be said for keeping the dependencies of the `api` module as few in number as possible, so I think we should leave it as-is for now. Good call.

I'll mark this comment thread as 'resolved'.

HansBrende:
@lewismc I can potentially do a quick fix-up on this PR this weekend when I get a chance, and then we should be good to go.

PR #162, opened 2020-10-07
ANY23-455 Format entire Any23 codebase with formatter-maven-plugin

lewismc:
This PR addresses https://issues.apache.org/jira/browse/ANY23-455

It is a large bulk change and was simply generated by using the `mvn formatter:format` target of the [formatter-maven-plugin](https://code.revelc.net/formatter-maven-plugin/usage.html)


lewismc:
We can use the `mvn formatter:validate` target to check the formatting.

PR #229, opened 2021-12-17
ANY23-536 Upgrade to tika 2.2.0

lewismc:
PR addresses https://issues.apache.org/jira/browse/ANY23-536


lewismc:
The dependency is not yet available in Maven Central.

PR #246, opened 2022-01-09
add edit source button in generated site

hboutemy:
Recent Fluido skins you are using can have updated breadcrumb that eases documentation edit: there is an "edit" icon that goes directly to page source on GitHub to provide a PR


lewismc:
Excellent thank you @hboutemy

